PR #91, opened 2020-04-10
Luma API update, animometer example revisited

ilijapuaca:
- Revisited some of the API in `Model`/`AnimationLoop`
- `animometer` example now uses the updated API instead of making calls to dawn API directly
- Minor dependency cleanup

Few things worth noting:
- There seems to be a memory leak in the run loop (it's there even in example provided by dawn, without any modifications)
- Most of the WebGPU objects are being passed around by value across the API. This initially seemed weird to me so I started trying to turn them into pointers, which makes the function calls somewhat sloppy. Any particular reason they are being passed by value, are they that lightweight? I assumed that making copies of these was causing the memory leak, based on this text from WebGPU spec:
> WebGPU sees physical GPU hardware as GPUAdapters. It provides a connection to an adapter via GPUDevice, which **manages resources**...
- Did we give threading some thought when it comes to rendering?


ilijapuaca:
Landing this together with #92, will address any comments as they come

PR #128, opened 2020-05-05
Added manhattan population example, Xcode project cleanup

ilijapuaca:
- Added another `ScatterplotLayer` example
- Added the missing `airports.ndjson` file to `CMakeLists`
- Cleaned up Xcode project, it now has separate targets for all of the examples so that they can be ran easily


ilijapuaca:
Merging this in order to avoid potential conflicts in the project file, will address any comments in a followup

PR #136, opened 2020-05-13
Added shaders for the SolidPolygonLayer

randypalusz:
#107 - Seems that shaders compile, although still getting ValidationErrors when running the vancouver-blocks example.


cpp/modules/deck.gl/layers/src/solid-polygon-layer/solid-polygon-layer-vertex-main.glsl.h
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  gl_Position = project_position_to_clipspace(pos, pos64Low, vec3(0.), geometry.position);
+
+  if (layerOptions.extruded) {
+    // TODO(randy@unfolded.ai) - Bypassing actual lightColor calculation for now, figure out why this is happening:
+    // libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Shader compilation failed with error: myshader?:281: error: 'lighting_getLightColor' : no matching overloaded function found
+    // myshader?:281: error: '=' :  cannot convert from ' const float' to ' temp highp 3-component vector of float'
+
+    // vec3 lightColor = lighting_getLightColor(colors.rgb, project.uCameraPosition, geometry.position.xyz, normal);
```
randypalusz, line 104: See the comment directly above this line. When using this, it seems that lighting_getLightColor cannot be called using the given parameters.d

cpp/modules/deck.gl/layers/src/solid-polygon-layer/solid-polygon-layer.cpp
```diff
@@ -224,3 +243,86 @@ auto SolidPolygonLayer::getLineColorData(const std::shared_ptr<arrow::Table>& ta
...
+        std::make_shared<garrow::Field>("fillColors", wgpu::VertexFormat::Float4),
+        std::make_shared<garrow::Field>("lineColors", wgpu::VertexFormat::Float4),
+        std::make_shared<garrow::Field>("pickingColors", wgpu::VertexFormat::Float3)};
+
+    auto attributeSchema = std::make_shared<lumagl::garrow::Schema>(attributeFields);
+
+    // TODO(randy@unfolded.ai): Although instanced fields don't seem to be used in the JS vertex-top shader,
+    //                          leaving as nullptr causes a segmentation fault. Using placeholder for now.
+    std::vector<std::shared_ptr<garrow::Field>> instancedFields{
+        std::make_shared<garrow::Field>("placeholder", wgpu::VertexFormat::Float)};
```
randypalusz, line 273: See comment above. Seems that setting this to `nullptr` causes a segmentation fault. In the JS code, `isInstanced`, passed to the Model constructor, is not set to 1/true. So it seems we don't want instanced parameters like what is in `if (extruded) {` below?

cpp/modules/deck.gl/layers/src/solid-polygon-layer/solid-polygon-layer-vertex-main.glsl.h
ibgreen: Have we ported the lighting shader module? If not that explains it.

cpp/modules/deck.gl/layers/src/solid-polygon-layer/solid-polygon-layer.cpp
ibgreen: I believe we are using two table here, one contains instanced attributes and one contains non-instanced attributes.

Some layers do not use instanced attributes. I think the polygon layer does not use instanced attributes.

You need to make sure the call to pipeline->Draw inside Model has 0 instance count parameter.

randypalusz:
> You need to make sure the call to pipeline->Draw inside Model has 0 instance count parameter.

Seems that `pass.Draw` inside the Model class draws a minimum of one instance in the case of no instance parameters being passed:
image

cpp/modules/deck.gl/layers/src/solid-polygon-layer/solid-polygon-layer.cpp
```diff
@@ -224,3 +243,86 @@ auto SolidPolygonLayer::getLineColorData(const std::shared_ptr<arrow::Table>& ta
...
+    // make new model using vsTop
+    // TODO(randy@unfolded.ai): Remove instanced fields, add (wireframe, isSideVertex) uniforms to top
+    std::vector<std::shared_ptr<garrow::Field>> attributeFields{
+        std::make_shared<garrow::Field>("vertexPositions", wgpu::VertexFormat::Float2),
+        std::make_shared<garrow::Field>("vertexValid", wgpu::VertexFormat::Float),
+        std::make_shared<garrow::Field>("positions", wgpu::VertexFormat::Float3),
+        std::make_shared<garrow::Field>("elevations", wgpu::VertexFormat::Float),
+        std::make_shared<garrow::Field>("fillColors", wgpu::VertexFormat::Float4),
+        std::make_shared<garrow::Field>("lineColors", wgpu::VertexFormat::Float4),
+        std::make_shared<garrow::Field>("pickingColors", wgpu::VertexFormat::Float3)};
```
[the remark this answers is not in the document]
randypalusz: pickingColors dropped in shaders + src

cpp/modules/deck.gl/layers/src/solid-polygon-layer/solid-polygon-layer.h
```diff
@@ -46,13 +46,15 @@ class SolidPolygonLayer : public Layer {
  protected:
   void initializeState() override;
   // auto getPickingInfo() override;
-  void updateState(const ChangeFlags&, const Layer::Props* oldProps) override;
+  void updateStateSPL(const ChangeFlags&, const SolidPolygonLayer::Props* oldProps);
```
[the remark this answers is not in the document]
randypalusz: Made this change initially as it seemed that the "filled" and "extruded" props used in determining whether the models need to be regenerated are exclusive to the SolidPolygonLayer.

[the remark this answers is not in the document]
ilijapuaca: We don't seem to be calling this from anywhere though as far as I can tell? You should still be able to get fields out of `SolidPolygonLayer::Props` by casting the `Layer::Props` superclass that is being passed here.

We already cast current layer properties in the implementation: `auto props = std::dynamic_pointer_cast<SolidPolygonLayer::Props>(this->props());`, we should do the same for `oldProps` and revert this change

PR #145, opened 2020-05-29
SolidPolygonLayer made functional

ilijapuaca:
There is definitely some room for optimization, but the core functionality should be in there.

Spent quite a while debugging an issue where running the executable produced by cmake/clang doesn't seem to render anything, while running the same setup with Xcode works just fine, with FPS maxed out. Could someone please try building and running the example in order to make sure it doesn't have something to do with my environment, as I'm out of ideas at this point.

I also had to revisit some of the dirty flag setting code in order to make sure we don't do too much wasteful work.

Screen Shot 2020-05-29 at 18 24 40


cpp/modules/deck.gl/core/src/arrow/arrow-mapper.cc
```diff
@@ -24,6 +24,29 @@
...
+    -> std::shared_ptr<arrow::Array> {
+  arrow::MemoryPool* pool = arrow::default_memory_pool();
+  arrow::BooleanBuilder builder{pool};
+
+  for (auto i = 0; i < table->num_rows(); ++i) {
+    // TODO(ilija@unfolded.ai): Move row object creation out of the loop
+    auto row = Row{table, i};
+    auto value = getValueFromRow(row);
+
+    if (!builder.Append(value).ok()) {
```
isaacbrodsky, line 37: Call `Reserve` on the builder before appending values? I assume otherwise this works like a regular array builder and needs to resize it's buffer.

ibgreen:
> Spent quite a while debugging an issue where running the executable produced by cmake/clang doesn't seem to render anything, while running the same setup with Xcode works just fine, with FPS maxed out. Could someone please try building and running the example in order to make sure it doesn't have something to do with my environment, as I'm out of ideas at this point.

As far as I could tell, the example has not been added to the xcode proj. I tried a clang build from this branch, and the vancouver example does indeed fail with an `Unsupported data type` exception, and a quick search points me to `arrow-utils.cc`

<details>
```
deck.gl-native (master)$ ./build/clang/cpp/examples/deck.gl/vancouver-blocks
2020-05-30 12:08:17.012 vancouver-blocks[30144:7053037] MTLDevice newLibraryWithSource => Error Domain=MTLLibraryErrorDomain Code=4 "Compilation succeeded with:

program_source:33:22: warning: unused variable 'geometry'
    FragmentGeometry geometry;
                     ^
" UserInfo={NSLocalizedDescription=Compilation succeeded with:

program_source:33:22: warning: unused variable 'geometry'
    FragmentGeometry geometry;
                     ^
}
2020-05-30 12:08:17.203 vancouver-blocks[30144:7053037] MTLDevice newLibraryWithSource => Error Domain=MTLLibraryErrorDomain Code=4 "Compilation succeeded with:

program_source:33:22: warning: unused variable 'geometry'
    FragmentGeometry geometry;
                     ^
" UserInfo={NSLocalizedDescription=Compilation succeeded with:

program_source:33:22: warning: unused variable 'geometry'
    FragmentGeometry geometry;
                     ^
}
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Unsupported data type
Abort trap: 6
```
</details>

ilijapuaca:
I've added a new schema for this example that you should see in the top left (screenshot below).

How did you get that output? When I run the same executable, I do not get a crash, but rather the window stays open with nothing being drawn, and I am not seeing any of those logs.

Screen Shot 2020-05-31 at 09 58 05

cpp/modules/deck.gl/core/src/arrow/row.h
```diff
@@ -58,40 +58,91 @@ class Row {
...
+    return vectorData;
+  }
+  template <typename T>
+  auto getVector3List(const std::string& columnName, const std::vector<mathgl::Vector3<T>>& defaultValue = {}) const
+      -> std::vector<mathgl::Vector3<T>> {
+    auto listData = this->getNestedListData<T>(columnName, {});
+    std::vector<mathgl::Vector3<T>> vectorData;
+    vectorData.reserve(listData.size());
+    for (const auto& data : listData) {
+      vectorData.push_back(mathgl::Vector3<T>{data.size() > 0 ? data[0] : 0, data.size() > 1 ? data[1] : 0,
```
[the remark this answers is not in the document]
ilijapuaca: These aren't nested though? Not sure if I'm missing something, but I'm just doing bounds checking for each of the components

cpp/modules/deck.gl/core/src/arrow/arrow-mapper.cc
ilijapuaca: Correct, wasn't sure what a reasonable chunk to reserve would be, so I postponed it until indexing was in place as the number of rows has decreased.

Added a conservative estimate before we start iterating

ilijapuaca:
Spent quite a while trying to get the example to run from terminal without success. The exact same executable runs fine from Xcode, while when running it manually from terminal it doesn't render anything.

I can see that the render loop is running, and the profiler shows that calls are being made to Metal API. This seems like a ghost hunt at this point, any ideas are welcome, I'll give this another shot if I think of any new things to try

PR #155, opened 2020-06-04
Fixed the polygon layer uniform alignment

ilijapuaca:
While updating our iOS examples, I figured that the issue wasn't only with runs from terminal, as I wasn't getting anything drawn in the sample iOS app. This was much easier to debug there as I could inspect the actual data.

I tried a few things in order to fix the alignment and keep the `bool` members, but I was not able to get it working consistently, even with use of `#pragma pack`. I switched these over to `uint32_t` as that's guaranteed not to fail regardless of the compiler

### DIFF
--- a/cpp/modules/deck.gl/layers/src/solid-polygon-layer/solid-polygon-layer.h
+++ b/cpp/modules/deck.gl/layers/src/solid-polygon-layer/solid-polygon-layer.h
@@ -94,11 +94,11 @@ class SolidPolygonLayer::Props : public Layer::Props {
 };
 
 /// The order of fields in this structure is crucial for it to be mapped to its GLSL counterpart properly.
-/// bool has a 4-byte alignment in GLSL.
+/// Using uint32_t in place of bool in order to avoid packing issues, even when specifying 4-byte alignment.
 /// https://learnopengl.com/Advanced-OpenGL/Advanced-GLSL
 struct SolidPolygonLayerUniforms {
-  alignas(4) bool extruded;
-  alignas(4) bool wireframe;
+  uint32_t extruded;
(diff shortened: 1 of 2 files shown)


cpp/modules/deck.gl/layers/src/solid-polygon-layer/solid-polygon-layer.h
```diff
@@ -94,11 +94,11 @@ class SolidPolygonLayer::Props : public Layer::Props {
 };
 
 /// The order of fields in this structure is crucial for it to be mapped to its GLSL counterpart properly.
-/// bool has a 4-byte alignment in GLSL.
+/// Using uint32_t in place of bool in order to avoid packing issues, even when specifying 4-byte alignment.
 /// https://learnopengl.com/Advanced-OpenGL/Advanced-GLSL
 struct SolidPolygonLayerUniforms {
-  alignas(4) bool extruded;
-  alignas(4) bool wireframe;
+  uint32_t extruded;
```
isaacbrodsky, line 100: Do we actually know what the size of `bool` was when it was failing? I would be concerned that the size would change unexpectedly between compilers. Using `uint32_t` is certainly safer.

ilijapuaca: The size didn't vary, it was packing that made the memory be structured differently. Even with `#pragma pack` directives, I couldn't get it to pack these fields into 4-byte chunks with added padding consistently.

Perhaps I could've made it work with a bit more effort, but without access to all the compilers/versions that this could be compiled with, it's hard to say whether it's a fail-proof implementation like this one is supposed to be

PR #157, opened 2020-06-08
Deck drawing options renamed

ilijapuaca:


ibgreen (commented):
Does any example need to be updated? Since we are just aliasing the type, I suppose the example will still compile if it uses AnimationLoop version.

ilijapuaca:
> Does any example need to be updated? Since we are just aliasing the type, I suppose the example will still compile if it uses AnimationLoop version.

Custom drawing options were only used in `texture-render` example, which I've updated

